PR #2, opened 2016-05-27
Avoid backslash line continuation, |=, simplify with or

cclauss:
A space to the right of a backslash would break your script yet it is invisible to the reader.
|= shortens long lines and executes faster.
Or will gracefully deal with other Falsey values such as: [], {}, 0, False, etc.

### DIFF
--- a/fibratus/controller.py
+++ b/fibratus/controller.py
@@ -56,22 +56,22 @@ class KTraceProps():
     def enable_kflags(self, syscall=False, cswitch=False):
         # enable the basic set of flags
         # for the kernel events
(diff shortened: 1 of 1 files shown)


rabbitstack:
Those inliners are really cool! Thanks for the PR. However, it would violate the pylint rule `line break before binary operator`.Can you fix that before I merge it with the base branch?

cclauss:
Appveyor seems to fail the CI tests.

rabbitstack:
Test are passing OK. For some strange reason the submission of coverage reports to coveralls.io is failing. I'll look at this later.

PR #4, opened 2016-05-27
Tabs to spaces

cclauss:
Convert tabs to 4 spaces to placate landscape.io


cclauss:
https://landscape.io/diff/360164/fixed

PR #24, opened 2017-10-31
Enhance ES indexing

DynaMc:
I don't have a test env yet but will try asap. From a glance this appears it should fix the current ES indexing.

### DIFF
--- a/fibratus/output/elasticsearch.py
+++ b/fibratus/output/elasticsearch.py
@@ -22,7 +22,6 @@ from datetime import datetime
 
 
 class ElasticsearchOutput(Output):
-    dt = datetime.now()
 
     def __init__(self, **kwargs):
         """Creates an instance of the Elasticsearch output adapter.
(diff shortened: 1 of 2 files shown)


DynaMc:
hmm, I'll have to wait to setup my env. The test is hard coded as well. Unless you can try @rabbitstack

rabbitstack:
@DynaMc thanks for the PR! Obviously, we would have to specify the daily index depending on some configuration property, and not hard coded. I would suggest `index_type` yaml attribute. In case it's equal to `daily` then we would build the index accordingly. Test case would be nice, so we can make sure nothing is broken after introducing the change.

rabbitstack:
@DynaMc, the support for daily indices is in the master now. Even though you'll have to build the binary manually to be able to use this functionality. There are these two properties:

```
index_type: daily
daily_index_format: %Y.%m.%d
```

The first of them indicates the type of the index. `daily_index_format` specifies the pattern for your index names.

PR #126, opened 2022-08-23
chore: Raw event parsing and performance optimizations

rabbitstack:
Raw event parsing directly accesses the event buffer and extracts all the parameters according to the event schema version. Along with raw event parsing which brings huge performance improvements, a series of other optimizations have been introduced. Also, the trace controller has been refactored to facilitate the addition of new event sources.


yusufozturk:
🍾🎊🎉

PR #140, opened 2022-11-29
feat: Add additional TCP fields and typo fix

yusufozturk:
* A typo issue is fixed
* It might be useful to show tcp start time and end time.

### DIFF
--- a/pkg/kevent/kparams/fields_windows.go
+++ b/pkg/kevent/kparams/fields_windows.go
@@ -151,9 +151,9 @@ const (
 	// NetSeqNum is the parameter name that represents the TCP sequence number.
 	NetSeqNum = "seqnum"
 	// NetStartTime is the parameter name that represents the TCP start time.
-	NetStartTime = "starttime"
+	NetStartTime = "start_time"
 	// NetEndTime is the parameter name that represents the TCP end time.
-	NetEndTime = "endtime"
+	NetEndTime = "end_time"
 	// NetConnID is the parameter name that represents a unique connection identifier.
 	NetConnID = "connid"
 	// NetL4Proto is the parameter name that identifies the Layer 4 protocol name.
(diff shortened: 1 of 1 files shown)


rabbitstack:
Hi @yusufozturk,

Thanks for your contribution!

I got a couple of comments.

It looks like this PR is merely declaring the field names. If your intent was to include these fields in the corresponding events, you should add the conversion logic [here](https://github.com/rabbitstack/fibratus/blob/286afff927e5bdd48123a0ef7bbab6046cab3ced/pkg/kevent/kparams/canonicalize.go#L96) for TDH parsing, and read the fields from the event buffer in case of [raw](https://github.com/rabbitstack/fibratus/blob/286afff927e5bdd48123a0ef7bbab6046cab3ced/pkg/kstream/kstreamc_windows.go#L907) parsing.
It is also important to adhere to the event parameter naming semantics. `starttime` and `endtime` should adopt the camel case format, `start_time` and `end_time` respectively.
Lastly, it would be nice to add filter fields for these new parameters so we can write detection rule that rely on them,

yusufozturk:
Hi @rabbitstack ,

Thanks for the help. Actually I was able to figure that out, thanks to your easy design. But I was planning to send them via another PR. I wanted to push this earlier because it doesn't require any additional tests or something and I will need start time and end time to calculate latencies in the future.

It's sad that we can't see SYN events via ETW. At least when I search on Microsoft documentation, I didn't find any information about it.

rabbitstack:
> But I was planning to send them via another PR. 

Fair enough. If you adjust the field names as I explained in the previous comment, this will be good to merge.

> It's sad that we can't see SYN events via ETW.

I hadn't had a chance to dive deep enough, but the [`Microsoft-Windows-TCPIP`](https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Parsers/Microsoft-Windows-TCPIP.cs) ETW provider may emit such events.

rabbitstack:
Merging it now. Thanks for your contribution!

PR #717, opened 2026-08-05
chore(ebpf): Add Linux CO-RE eBPF feasibility spike

mostafa:

Validates that Fibratus can load CO-RE eBPF programs on Linux with the hard runtime contract needed for a future eBPF event source: kernel >=5.9, usable `/sys/kernel/btf/vmlinux`, ring buffers, and `iter/task` process baseline capture without `bpf_d_path`. The spike also proves shared-map replacement across separately generated objects, best-effort `/proc` enrichment, and race-safe startup reconciliation keyed by PID + start boot time. A feasibility note audits Windows coupling and every `ps.Snapshotter` consumer that must move before a Linux build can land.

---

/kind improvement

---

/area instrumentation

/area deps

/area build

---

- Throwaway spike lives under `internal/ebpf/spike` with committed bpf2go artifacts. Ordinary builds do not require clang; regenerate with `internal/ebpf/spike/generate.sh`.
- Runtime probe helpers are in `internal/ebpf.ProbePrerequisites`. `iter/task` attach remains the conclusive iterator check inside the spike runner.
- Validated in a privileged Docker Desktop LinuxKit 6.12 container with kernel BTF and tracefs mounted. Example successful run showed pending hot events queued during baseline and replayed (`pending_queued`/`replay_applied` > 0).
- No Fibratus CLI/runtime behavior changes on Windows.

How to reproduce the spike:

```bash
# privileged Linux host/container with /sys/kernel/btf/vmlinux and tracefs
go test ./internal/ebpf -count=1
SPIKE_DURATION=1s go run ./internal/ebpf/spike/cmd/spike
```

---

No.


rabbitstack:
@mostafa I ran the example you described in the pull request. All requirements for loading and executing ebpf programs were satisfied:

```bash
SPIKE_DURATION=1s sudo go run ./internal/ebpf/spike/cmd/spike
go: downloading github.com/cilium/ebpf v0.22.0
go: downloading golang.org/x/sys v0.43.0
kernel=7.1.5-arch1-1 btf=true ringbuf=true iter=true
snapshots=217 exec_events=80 reconciled=296
metrics: pending_queued=42 pending_dropped=0 ringbuf_dropped=0 enrichment_miss=133 replay_applied=42 snapshot_upserts=217
```

internal/ebpf/spike/c/execve.bpf.c
```diff
@@ -0,0 +1,53 @@
+//go:build ignore
+
+#include "events.h"
+#include "bpf_core_read.h"
+#include "bpf_tracing.h"
+
+char LICENSE[] SEC("license") = "Dual MIT/GPL";
+
+SEC("tp/sched/sched_process_exec")
```
rabbitstack, line 9: I understand this is a throwaway material, but I'm wondering about the hook attachment strategy. Is there any preference between sched_process_exec or a consistent syscall exit raw tracepoints?
The vast majority of events will share a header with common attributes, so maybe we may keep this in mind for the future work.

internal/ebpf/spike/run.go
```diff
@@ -0,0 +1,295 @@
...
+	"io"
+	"os"
+	"sync/atomic"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+	"github.com/cilium/ebpf/ringbuf"
+	"github.com/cilium/ebpf/rlimit"
+	fibebpf "github.com/rabbitstack/fibratus/internal/ebpf"
```
rabbitstack:
```suggestion
	libebpf "github.com/rabbitstack/fibratus/internal/ebpf"
```

internal/ebpf/spike/execve_bpfel.go
```diff
@@ -0,0 +1,136 @@
...
+
+package spike
+
+import (
+	"bytes"
+	_ "embed"
+	"fmt"
+	"io"
+
+	"github.com/cilium/ebpf"
```
rabbitstack, line 12: Linter is complaining about missing package. It seems like it wasn't added to go.mod

rabbitstack (changes requested):

internal/ebpf/spike/c/execve.bpf.c
mostafa: Good call. For production I prefer consistent raw syscall entry/exit tracepoints so the vast majority of events can share one common header (pid/tgid/uid/comm/timestamp/syscall id/return), with per-CPU scratch for entry/exit correlation. `sched_process_exec` is convenient in the spike for successful-exec filename, but I would treat it as complementary rather than the default attach strategy going forward.

internal/ebpf/spike/run.go
mostafa: Done in 2317f13. Renamed the import alias to `libebpf`.

internal/ebpf/spike/execve_bpfel.go
mostafa: Good catch. `github.com/cilium/ebpf` was lost from `go.mod` during the rebase onto `linux-port` (Darwin `go mod tidy` drops Linux-only imports). Restored it as a direct require in 2317f13 and noted to tidy with `GOOS=linux` going forward.

mostafa:
Thanks for running the spike on Arch and for the review. Addressed the change requests in 2317f13:

- restored `github.com/cilium/ebpf` as a direct `go.mod` require
- renamed the import alias to `libebpf`
- replied on the attach strategy: prefer raw syscall entry/exit with a shared header for production; keep `sched_process_exec` as complementary for successful-exec filename

Also aligned with using `linux-port` as the integration base for follow-up PRs.

rabbitstack (approved):
LGTM! Thanks for your contribution.

PR #721, opened 2026-08-08
refactor(bootstrap): Add Linux build and domain foundations

mostafa:

Makes `GOOS=linux go build ./cmd/fibratus` succeed and establishes the Linux domain foundations for the eBPF event source. It splits platform-specific bootstrap and CLI entrypoints, defines stable Linux event semantics and parameter catalogs, introduces Linux event-source configuration, and provides an in-memory process snapshotter with Linux-native 64-bit PIDs.

Shared configuration, event formatting/marshalling, filtering, rule compilation, alert serialization, and utility behavior now use explicit Linux/Windows implementations where their models differ. Windows-only packages/tests remain gated so the Linux dependency graph does not pull `golang.org/x/sys/windows`.

---

/kind refactor

/kind feature

---

/area build

/area event

/area config

/area telemetry

/area rule-engine

/area filters

/area cli

---

- The Linux CLI path includes bootstrap and `EventSource` implementations. `Run` returns `ErrCaptureNotWired`, so this change does not provide live Linux capture.
- Linux `event.Type` values are project-owned semantic IDs (`uint16`), not syscall numbers.
- `event.PID` is platform-specific: `uint64` on Linux and `uint32` on Windows. Shared consumers use each platform's `ps.Snapshotter` interface.
- Linux process/thread creation semantics distinguish `clone` events using `CloneFlags` and `CLONE_THREAD`.
- Event types, parameter names/types, field definitions, filter functions, compiler behavior, queues, formatting, marshalling, and alert process serialization are split by platform where required.
- Common configuration, output loading, transformers, and schema validation are shared. The schema accepts Linux event-source flags and Linux event names.
- Windows behavior remains covered by the full Windows package compilation suite.
- Verified:
  - Docker Linux: `go build -o /tmp/fibratus ./cmd/fibratus`
  - Docker Linux: `go test ./pkg/ps ./pkg/event ./pkg/config ./pkg/rules ./pkg/filter ./internal/bootstrap`
  - `GOOS=windows GOARCH=amd64 go build -o /tmp/fibratus.exe ./cmd/fibratus`
  - `GOOS=windows GOARCH=amd64 go test -exec=true ./...`

---

No live Linux capture is introduced. The Linux binary builds and exposes the CLI surface, but event capture is not wired.


cmd/fibratus/app/config/config.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: As far as I can see, there is nothing platform-specific in this source file. Ideally, we should gate the `Config` struct for the Linux build (`rest.Get`) shouldn't require a big rework (maybe only handling the named pipe transport).

pkg/event/pid_linux.go
```diff
@@ -0,0 +1,25 @@
+//go:build linux
```
rabbitstack, line 1: Let's rollback this change, as per above comment. The `config` command structure is platform agnostic. Just the pieces it imports shall be adapted to Linux.

cmd/fibratus/app/list/list.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Essentially, the same comment as in `config.go`. `fields.Get`, `event.GetTypesMeta` can be noop in Linux until we bring the full support.

pkg/config/alertsender_linux.go
```diff
@@ -0,0 +1,25 @@
+//go:build linux
```
rabbitstack, line 1: The `list` command is platform-independent, so, let's rollback this change.

cmd/fibratus/app/rules/create.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Rules management CLI is platform agnostic.

cmd/fibratus/app/rules/list.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Rules management CLI is platform agnostic.

cmd/fibratus/app/rules/rules.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Rules management CLI is platform agnostic.

pkg/api/listener_linux.go
```diff
@@ -0,0 +1,25 @@
+//go:build linux
```
rabbitstack, line 1: Let's rollback, same reason as the previous source files.

cmd/fibratus/app/rules/validate.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Rules management CLI is platform agnostic.

cmd/fibratus/app/stats/stats.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Just the `Stats` structure will change to accomodate Linux specifics. The `stats` command entrypoint is shared accross platforms.

cmd/fibratus/app/stats/stats_linux.go
```diff
@@ -0,0 +1,25 @@
+//go:build linux
```
rabbitstack, line 1: Let's just declare the `Stats` structure here. Can initially start with no fields. Also, let`s move the current `Stats` structure to `stats_windows.go` source file.

cmd/fibratus/main_linux.go
```diff
@@ -0,0 +1,32 @@
...
+package main
+
+import (
+	"os"
+
+	"github.com/rabbitstack/fibratus/cmd/fibratus/app"
+)
+
+func main() {
+	if err := app.RootCmd.Execute(); err != nil {
```
rabbitstack, line 29: Are you planning to abstract away this the same in Windows where the `App` component does the heavy lifting?

P.S. I just saw that you already prepared the `App` counterpart for Linux. Let's consume it then and align with the Windows implementation.

internal/bootstrap/bootstrap_linux.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"errors"
+
+	"github.com/rabbitstack/fibratus/pkg/config"
+)
+
+var ErrCaptureNotWired = errors.New("linux event capture is not wired yet")
+
+type App struct {
```
rabbitstack, line 32: I imagine you'll extend/complete the functions in upcoming pull requests?

internal/bootstrap/bootstrap_windows.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack, line 1: Let's rename the source file to `bootstrap_windows.go`, rather the use the build tag, for the sake of consistency.

internal/bootstrap/source_windows.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack, line 1: Same here. Prefer `source_windows.go` as we have `source_linux.go`

pkg/aggregator/aggregator_windows.go
```diff
@@ -0,0 +1,28 @@
+//go:build windows
```
rabbitstack, line 1: I'd refactor to `aggregator_windows.go`. If ever the need arises to have something specific for Linux, we can have `aggregator_linux.go`, while the base logic is in `aggregator.go`.

pkg/callstack/callstack_windows.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack, line 1: Let's use the file suffix. Rename to `callstack_windows.go`

pkg/cap/version/version_linux.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+const (
+	ProcessSecV1 Version = iota + 1
+	ProcessSecV2
+	ProcessSecV3
+	ProcessSecV4
+)
+
+const (
```
rabbitstack: Handle section shall not exist in Linux.

pkg/cap/version/version_linux.go
```diff
@@ -0,0 +1,45 @@
...
+	HandleSecV1 Version = iota + 1
+)
+
+const (
```
rabbitstack: PE section shall not exist in Linux.

pkg/ps/resolver.go
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ps
+
+import pstypes "github.com/rabbitstack/fibratus/pkg/ps/types"
+
+type Resolver interface {
+	Find(pid uint32) (bool, *pstypes.PS)
```
rabbitstack: The pid/tid in Linux is represented as an unsigned 64 bits integer. So, we'll need an interface with the generic parameter:

```go
type Resolver[P any] interface {
	Find(pid P) (bool, *pstypes.PS)
}
```

Then embed in Linux and Windows snapshotter interfaces.

snapshotter_windows.go:

```go
type Snapshotter interface {
  Resolver[uint32]
}
```

snapshotter_linux.go:

```go
type Snapshotter interface {
  Resolver[uint64]
}
```

rabbitstack: I'd prefer to have this source named as `snapshotter.go` with the generic `Resolver` interface. This interface could also contain other operations that are shared between Windows and Linux. For example.

```go
// FindAndPut attempts to retrieve process' state for the specified process identifier.
	// If the process is found, the snapshotter state is updated with the new process.
	FindAndPut(pid P) *pstypes.PS
	// Put inserts the process state into snapshotter.
	Put(*pstypes.PS)
	// Size returns the total number of process state items.
	Size() uint32
	// Close closes process snapshotter and disposes all allocated resources.
	Close() error
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"sync"
+
+	"github.com/rabbitstack/fibratus/pkg/event"
+	pstypes "github.com/rabbitstack/fibratus/pkg/ps/types"
+	"github.com/rabbitstack/fibratus/pkg/util/va"
+)
+
+// LinuxSnapshotter stores Linux process state for enrichment and rule evaluation.
+type LinuxSnapshotter interface {
```
rabbitstack:
```suggestion
type Snapshotter interface {
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"sync"
+
+	"github.com/rabbitstack/fibratus/pkg/event"
+	pstypes "github.com/rabbitstack/fibratus/pkg/ps/types"
+	"github.com/rabbitstack/fibratus/pkg/util/va"
+)
+
+// LinuxSnapshotter stores Linux process state for enrichment and rule evaluation.
```
rabbitstack:
```suggestion
// Snapshotter stores Linux process state for enrichment and rule evaluation.
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	Put(*pstypes.PS)
+	Size() uint32
+	Close() error
+	AddThread(*event.Event) error
+	RemoveThread(pid uint32, tid uint32) error
+	AddMmap(*event.Event) error
+	RemoveMmap(pid uint32, addr va.Address) error
+}
+
+type linuxSnapshotter struct {
```
rabbitstack:
```suggestion
type snapshotter struct {
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	mu    sync.RWMutex
+	procs map[uint32]*pstypes.PS
```
rabbitstack:
```suggestion
	procs map[uint64]*pstypes.PS
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// NewLinuxSnapshotter builds an in-memory Linux process snapshotter.
+func NewLinuxSnapshotter() LinuxSnapshotter {
```
rabbitstack:
```suggestion
func NewSnapshotter() Snapshotter {
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	AddMmap(*event.Event) error
+	RemoveMmap(pid uint32, addr va.Address) error
+}
+
+type linuxSnapshotter struct {
+	mu    sync.RWMutex
+	procs map[uint32]*pstypes.PS
+}
+
+// NewLinuxSnapshotter builds an in-memory Linux process snapshotter.
```
rabbitstack:
```suggestion
// NewSnapshotter builds an in-memory Linux process snapshotter.
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+func NewLinuxSnapshotter() LinuxSnapshotter {
+	return &linuxSnapshotter{procs: make(map[uint32]*pstypes.PS)}
```
rabbitstack:
```suggestion
	return &snapshotter{procs: make(map[uint64]*pstypes.PS)}
```

pkg/ps/resolver.go
rabbitstack: @mostafa I've been ruminating, and I think we could avoid the interface composition by giving a separate `Snapshotter` interface per platform. WDYT?

pkg/config/alertsender.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: It would be worth segregating per platform. (`alertsender_windows.go` and `alertsender_linux.go`), with `tryLoadAlertSenders` refactored and colocated in the generic `alertsender.go` source file in a way that it can handle (mail and Slack alertsenders, which are platform-independent), so that we can load the rest of the alertsenders that are relevant for Windows (systray, eventlog). Leaving the design decision to you.

pkg/config/config_linux.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/rabbitstack/fibratus/pkg/util/log"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+const configFile = "config-file"
+
+type Config struct {
```
rabbitstack: Apparently, most of the code could live in a platform-independent source file (`config.go`). Again, I'd divide it between Windows and Linux (`config_windows.go`, `config_linux.go`, and the shared `config.go`)

pkg/config/eventsource_linux.go
```diff
@@ -0,0 +1,101 @@
...
+const (
+	enableProcessEvents = "eventsource.enable-process"
+	enableFileEvents    = "eventsource.enable-file"
+	enableNetEvents     = "eventsource.enable-net"
+	enableMemEvents     = "eventsource.enable-mem"
+	excludedEvents      = "eventsource.blacklist.events"
+	excludedImages      = "eventsource.blacklist.images"
+)
+
+type EventSourceConfig struct {
```
rabbitstack, line 37: Another candidate to be shared between Windows and Linux. The only nuance is `event.AllWithState` function, which doesn't apply to Linux.

pkg/config/output.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Everything inside `output.go` can be reused in Linux, except the `eventlog` output and the Windows service check. So, let's create `output_windows.go` and `output_linux.go`, with anything shared between them in `output.go`

pkg/config/output_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Same for tests - Linux and Windows variants.

pkg/config/print_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: No need to gate; it should run on both platforms.

pkg/config/schema_linux.go
```diff
@@ -0,0 +1,32 @@
+//go:build linux
```
rabbitstack: This is the clone `schema_windows.go`. Instead of replicating, I'd remove the file suffix. (`schema_windows.go` -> `schema.go`)

pkg/config/transformer.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Transformers don't contain anything OS-specific. Should run on Windows and Linux.

pkg/config/transformer_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Same here.

pkg/event/params/names.go
rabbitstack, line 1: We really need to retain per-platform param names/types. Linux and Windows will diverge here. It would be worth keeping the shared ones in platform-independent source files.

pkg/event/params/param_types.go
rabbitstack, line 1: We really need to retain per-platform param names/types. Linux and Windows will diverge here. It would be worth keeping the shared ones in platform-independent source files.

pkg/event/batch_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Let's make sure we also exercise the Linux tests. Add the `windows` suffix.

pkg/event/bitset_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Let's make sure we also exercise the Linux tests. Add the `windows` suffix.

pkg/event/enum.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Use the file suffix constraint rather than the build tag.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	capver "github.com/rabbitstack/fibratus/pkg/cap/version"
+)
+
+func (e *Event) IsCreateFile() bool             { return false }
+func (e *Event) IsCreateProcess() bool          { return e.Type == CreateProcess || e.Type == Execve || e.Type == Clone }
+func (e *Event) IsCreateProcessInternal() bool  { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsCreateThread() bool           { return false }
+func (e *Event) IsCloseFile() bool              { return false }
+func (e *Event) IsCreateHandle() bool           { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsCloseHandle() bool            { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsDeleteFile() bool             { return false }
+func (e *Event) IsRenameFile() bool             { return false }
+func (e *Event) IsEnumDirectory() bool          { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsTerminateProcess() bool       { return e.Type == TerminateProcess || e.Type == Exit }
+func (e *Event) IsTerminateThread() bool        { return false }
+func (e *Event) IsUnloadModule() bool           { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsLoadModule() bool             { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsLoadModuleInternal() bool     { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsModuleRundown() bool          { return false }
```
rabbitstack: Not applicable to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsFileOpEnd() bool              { return false }
```
rabbitstack: Line 46-57 do not apply to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsVirtualAlloc() bool           { return false }
+func (e *Event) IsMapViewFile() bool            { return false }
+func (e *Event) IsUnmapViewFile() bool          { return false }
+func (e *Event) IsStackWalk() bool              { return false }
+func (e *Event) IsOpenThread() bool             { return false }
+func (e *Event) IsOpenProcess() bool            { return false }
+func (e *Event) IsNetworkTCP() bool             { return false }
+func (e *Event) IsNetworkUDP() bool             { return false }
+func (e *Event) IsDNS() bool                    { return false }
+func (e *Event) IsRundown() bool                { return e.IsProcessRundown() }
```
rabbitstack: Does not apply to Linux

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsSuccess() bool                { return true }
+func (e *Event) IsRundownProcessed() bool       { return false }
```
rabbitstack: Line 63-68 don't apply to Linux

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsState() bool                  { return e.Type.OnlyState() }
+
+// PartialKey returns a key used to deduplicate sequence partials.
+// On Linux this combines the event type with the process identifier.
+func (e *Event) PartialKey() uint64 {
+	return uint64(e.Type)<<32 | uint64(e.PID)
+}
+
+// HookID returns the stable bitset index for this type.
+func (t Type) HookID() uint16 { return uint16(t) }
```
rabbitstack: Does not apply to Linux.

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Event) IsOpenDisposition() bool        { return false }
+func (e *Event) IsCreateRemoteThread() bool     { return false }
+func (e *Event) IsSurrogateProcess() bool       { return false }
+func (e *Event) IsSystemPid() bool              { return e.PID == 0 || e.PID == 1 }
+func (e *Event) IsState() bool                  { return e.Type.OnlyState() }
+
+// PartialKey returns a key used to deduplicate sequence partials.
+// On Linux this combines the event type with the process identifier.
+func (e *Event) PartialKey() uint64 {
+	return uint64(e.Type)<<32 | uint64(e.PID)
```
rabbitstack: We can keep it as this for now, but this method should contemplate the event parameters to compose the partial key. However, I'm reworking the sequence engine to abolish partial key evaluations (but it will require some time).

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// HookID returns the stable bitset index for this type.
+func (t Type) HookID() uint16 { return uint16(t) }
+
+// MarshalJSON serializes the event to JSON.
+func (e *Event) MarshalJSON() []byte {
```
rabbitstack: I'd move to `marshaller_linux.go`

pkg/event/event_linux.go
```diff
@@ -0,0 +1,94 @@
...
+	b, err := json.Marshal(e)
+	if err != nil {
+		return []byte("{}")
+	}
+	return b
+}
+
+// UnmarshalRaw recovers the event from a capture buffer.
+// Captures are not wired on Linux yet.
+func (e *Event) UnmarshalRaw(_ []byte, _ capver.Version) error {
```
rabbitstack: I'd move to `marshaller_linux.go`

pkg/event/flags.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Add the file suffix constraint -> `flags_windows.go`

pkg/event/flags_linux.go
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package event
+
+import "strings"
+
+type ParamFlag struct {
```
rabbitstack: Move all the code to `flags.go` which is shared between Linux and Windows.

pkg/event/flags_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: It is better if we can also run these tests on Linux too.

pkg/event/formatter_linux.go
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+package event
+
+import "strconv"
+
+// Format applies the template on the provided event.
+func (f *Formatter) Format(evt *Event) []byte {
+	if evt == nil {
```
rabbitstack: This is essentially the same as the Windows counterpart. Maybe we can move to a shared `format` method?

pkg/event/formatter_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: I'm inclined to run these tests on Linux as well.

pkg/event/marshaller_linux.go
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package event
+
+var (
+	// SerializeHandles indicates if handles are serialized as part of the process state
+	SerializeHandles bool
+	// SerializeThreads indicates if threads are serialized as part of the process state
+	SerializeThreads bool
```
rabbitstack: This is the only flag applicable to Linux. We should remove the remaining ones.

pkg/event/marshaller_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Can we run it on Linux too?

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+import "slices"
+
+type Info struct {
+	Name        string
+	Category    Category
+	Description string
+}
+
+var events = map[Type]Info{
+	CreateProcess:    {Name: "CreateProcess", Category: Process, Description: "Creates a new process"},
```
rabbitstack: What is the difference between CreateProcess and Exceve. Or CreateProcess and Clone? I'd get rid of `CreateProcess`, `TerminateProcess`, and `ProcessRundown` as they are the direct incarnation of Windows API/syscall nomenclature. `Execve`, `Clone`, and `Exit` are much more consistent with the Linux API domain.

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	Execve:           {Name: "Execve", Category: Process, Description: "Executes a program"},
```
rabbitstack:
```suggestion
	Execve:           {Name: "execve", Category: Process, Description: "Executes a program"},
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	TerminateProcess: {Name: "TerminateProcess", Category: Process, Description: "Terminates a process"},
+	Exit:             {Name: "Exit", Category: Process, Description: "Exits a process"},
```
rabbitstack:
```suggestion
	Exit:             {Name: "exit", Category: Process, Description: "Exit all threads in a process"},
```

Just a note on the eBPF probe side: we'll need to capture the `sys_exit_group` syscall.

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	ProcessRundown:   {Name: "ProcessRundown", Category: Process, Description: "Reports an existing process"},
+	Clone:            {Name: "Clone", Category: Process, Description: "Clones a process or thread"},
```
rabbitstack:
```suggestion
	Clone:            {Name: "clone", Category: Process, Description: "Creates a child process or a thread"},
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+var types = map[string]Type{
+	"CreateProcess":    CreateProcess,
```
rabbitstack: Remove

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	"Execve":           Execve,
```
rabbitstack:
```suggestion
	"execve":           Execve,
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	"TerminateProcess": TerminateProcess,
```
rabbitstack: Remove

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	"Exit":             Exit,
```
rabbitstack:
```suggestion
	"exit":             Exit,
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	"ProcessRundown":   ProcessRundown,
```
rabbitstack: Remove

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+	"Clone":            Clone,
```
rabbitstack:
```suggestion
	"clone":            Clone,
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func All() []Type {
+	return []Type{CreateProcess, Execve, TerminateProcess, Exit, Clone}
```
rabbitstack:
```suggestion
	return []Type{Execve, Exit, Clone}
```

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func AllWithState() []Type {
```
rabbitstack: I think this is not going to be relevant on Linux.

pkg/event/metainfo_linux.go
```diff
@@ -0,0 +1,108 @@
...
+		return typ
+	}
+	return UnknownType
+}
+
+func NameToTypes(name string) []Type {
+	return []Type{NameToType(name)}
+}
+
+func GetTypesMeta() []Info {
```
rabbitstack: Can be placed in a shared source file (`metainfo.go`)

pkg/event/batch_linux_test.go
rabbitstack, line 1: I think we need to keep per-platform parameter definitions with the shared code living in `param.go`. So, rename this to `param_linux.go`.

pkg/event/param_capture_windows.go
rabbitstack, line 1: Rename to `param_windows.go` and move all the code from `param.go` that does not apply to Linux here.

pkg/event/param_linux.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+	"time"
+
+	"github.com/rabbitstack/fibratus/pkg/event/params"
+	"github.com/rabbitstack/fibratus/pkg/util/ntstatus"
+	"github.com/rabbitstack/fibratus/pkg/util/va"
+)
+
+// NewParam creates a new event parameter.
+func NewParam(name string, typ params.Type, value params.Value, options ...ParamOption) *Param {
```
rabbitstack: This can probably live in the shared `param.go` source file.

pkg/event/param_linux.go
```diff
@@ -0,0 +1,125 @@
...
+	return &Param{Name: name, Type: typ, Value: value, Flags: opts.flags, Enum: opts.enum}
+}
+
+// String returns the string representation of the parameter value.
+func (p Param) String() string {
+	if p.Value == nil {
+		return ""
+	}
+	switch p.Type {
+	case params.UnicodeString, params.AnsiString, params.Path, params.DOSPath, params.Key, params.HandleType:
```
rabbitstack: In Linux, all strings are UTF-8 encoded. So, we need to create a new `params.String` in `params/types_linux.go`.
DOSPath, Key, and HandleType are Windows-specific and should be removed.

pkg/event/param_linux.go
```diff
@@ -0,0 +1,125 @@
...
+		if s, ok := p.Value.(string); ok {
+			return s
+		}
+		return fmt.Sprintf("%v", p.Value)
+	case params.Status:
+		v, ok := p.Value.(uint32)
+		if !ok {
+			return ""
+		}
+		return ntstatus.FormatMessage(v)
```
rabbitstack: Windows status code to error message resolution. Doesn't apply here.

pkg/event/queue_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Can the tests be run on Linux?

pkg/event/formatter_linux_test.go
```diff
@@ -0,0 +1,36 @@
+//go:build linux
```
rabbitstack, line 1: I'd rather remove it than create a stub for a concept that we'll never have in the Linux implementation.

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"github.com/rabbitstack/fibratus/pkg/util/colorizer"
+	"github.com/rabbitstack/fibratus/pkg/util/hashers"
+)
+
+type Source uint8
+
+const (
+	SystemLogger Source = iota
```
rabbitstack: The source should be something like `RawSyscallTracepoint`

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	SecurityTelemetryLogger
+)
+
+type Type uint16
+
+const (
+	UnknownType Type = iota
+	CreateProcess
```
rabbitstack: Remove.

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	Execve
+	TerminateProcess
```
rabbitstack: Remove

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	Exit
+	ProcessRundown
```
rabbitstack: Remove

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	Clone
+
+	// Windows-only event types stubbed so shared consumers (rules compiler)
```
rabbitstack: Please, let's remove all of them and adapt the rule engine for the Linux-specific bits.

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	SetFileInformation
+	CreateFile
+	MapViewFile
+	UnmapViewFile
+	SetThreadContext
+	CreateSymbolicLinkObject
+	CreateThread
+)
+
+func (t Type) String() string {
```
rabbitstack, line 47: It is better to avoid map access in the hot path. Can we use the `switch` statement like in Windows?

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	return TypeToEventInfo(t).Name
+}
+
+func (t Type) Category() Category {
```
rabbitstack, line 61: It is better to avoid map access in the hot path. Can we use the `switch` statement like in Windows?

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	return TypeToEventInfo(t).Category
+}
+
+func (t Type) Subcategory() Subcategory {
+	return None
+}
+
+func (t Type) Description() string {
+	return TypeToEventInfo(t).Description
```
rabbitstack: It is better to avoid map access in the hot path. Can we use the `switch` statement like in Windows?

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+		return 0
+	}
+	return hashers.FnvUint32([]byte(t.String()))
+}
+
+func (t Type) Exists() bool {
+	return t != UnknownType && t.String() != "N/A"
+}
+
+func (t Type) OnlyState() bool {
```
rabbitstack: Does not apply to Linux.

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	return t == ProcessRundown
+}
+
+func (t Type) CanEnrichStack() bool {
```
rabbitstack: Does not apply to Linux.

pkg/event/types_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	*t = NameToType(name)
+	return nil
+}
+
+func (t Type) ID() uint {
+	return uint(t)
+}
+
+func (t Type) Source() Source {
+	return SystemLogger
```
rabbitstack: Adjust as per the previous comment.

pkg/filter/fields/defs.go
rabbitstack, line 1: Same as with parameter names/types. We have to keep per-platform definitions, with the common denominator in the shared source file. For example, most of the `evt.*` and `ps.*` can be shared between Linux and Windows.

pkg/config/output_linux.go
```diff
@@ -0,0 +1,29 @@
+//go:build linux
```
rabbitstack, line 1: It won't ever happen on Linux :).

pkg/filter/ql/functions/minidump_linux.go
```diff
@@ -0,0 +1,39 @@
+//go:build linux
```
rabbitstack: Does not apply to Linux.

pkg/filter/ql/function_linux.go
rabbitstack, line 1: This should be named `function_linux.go`

pkg/filter/ql/function_windows.go
rabbitstack, line 1: This should be named `function_windows.go`. Here you'll have to move the registration of minidump/get_reg_value functions, while all the rest are kept in `function.go`

pkg/filter/accessor_linux.go
```diff
@@ -0,0 +1,117 @@
...
+		if e.PS == nil {
+			return nil, ErrPsNil
+		}
+		return e.PS.Ppid, nil
+	case fields.PsName:
+		if e.PS == nil {
+			return nil, ErrPsNil
+		}
+		return e.PS.Name, nil
+	case fields.PsComm, fields.PsCmdline:
```
rabbitstack: PsComm is a deprecated field. Remove it.

pkg/filter/accessor_linux.go
```diff
@@ -0,0 +1,117 @@
...
+	psnap ps.Resolver
+}
+
+func (psAccessor) SetFields([]Field)            {}
+func (psAccessor) SetSegments([]fields.Segment) {}
+func (psAccessor) IsFieldAccessible(e *event.Event) bool {
+	return e.PS != nil || e.Category == event.Process
+}
+
+func newPSAccessor(psnap ps.Resolver) Accessor { return &psAccessor{psnap: psnap} }
```
rabbitstack:
```suggestion
func newPSAccessor(psnap ps.Snapshotter) Accessor { return &psAccessor{psnap: psnap} }
```

pkg/filter/accessor_linux.go
```diff
@@ -0,0 +1,117 @@
...
+// GetAccessors returns the accessors available on Linux.
+func GetAccessors() []Accessor {
+	return []Accessor{
+		newEventAccessor(),
+		newPSAccessor(nil),
+	}
+}
+
+type psAccessor struct {
+	psnap ps.Resolver
```
rabbitstack:
```suggestion
	psnap ps.Snapshotter
```

pkg/config/output_linux_test.go
rabbitstack, line 1: Rename to `accessor_linux.go`

pkg/filter/accessor_narrow_windows.go
```diff
@@ -0,0 +1,105 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package filter
+
+func (f *filter) pruneUnusedAccessors() {
```
rabbitstack: Move to `accessor_windows.go` and remove `accessor_narrow_windows.go`.

pkg/filter/accessor_windows.go
```diff
@@ -72,7 +72,7 @@ func getParentPs(e *event.Event) *pstypes.PS {
 
 // psAccessor extracts process's state or event specific values.
 type psAccessor struct {
-	psnap psnap.Snapshotter
+	psnap psnap.Resolver
```
rabbitstack:
```suggestion
	psnap psnap.Snapshotter
```

pkg/filter/accessor_windows.go
```diff
@@ -81,7 +81,7 @@ func (psAccessor) IsFieldAccessible(e *event.Event) bool {
 	return e.PS != nil || e.Category == event.Process
 }
 
-func newPSAccessor(psnap psnap.Snapshotter) Accessor { return &psAccessor{psnap: psnap} }
+func newPSAccessor(psnap psnap.Resolver) Accessor { return &psAccessor{psnap: psnap} }
```
rabbitstack:
```suggestion
func newPSAccessor(psnap psnap.Snapshotter) Accessor { return &psAccessor{psnap: psnap} }
```

pkg/filter/filter_linux.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+
+	"github.com/rabbitstack/fibratus/pkg/config"
+	"github.com/rabbitstack/fibratus/pkg/filter/fields"
+	"github.com/rabbitstack/fibratus/pkg/filter/ql"
+	"github.com/rabbitstack/fibratus/pkg/ps"
+)
+
+type opts struct {
+	psnap ps.Resolver
```
rabbitstack:
```suggestion
	psnap ps.Snapshotter
```

pkg/filter/filter_linux.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Option defines the option supplied to the filter
+type Option func(o *opts)
+
+// WithPSnapshotter passes a process resolver reference to the filter.
+func WithPSnapshotter(psnap ps.Resolver) Option {
```
rabbitstack:
```suggestion
func WithPSnapshotter(psnap ps.Snapshotter) Option {
```

pkg/filter/filter_linux.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type opts struct {
+	psnap ps.Resolver
+}
+
+// Option defines the option supplied to the filter
+type Option func(o *opts)
+
+// WithPSnapshotter passes a process resolver reference to the filter.
```
rabbitstack:
```suggestion
// WithPSnapshotter passes a process snapshotter reference to the filter.
```

pkg/filter/filter_windows_test.go
rabbitstack, line 1: Rename the file to `filter_test_windows.go`

pkg/filter/accessor_stubs_linux.go
```diff
@@ -0,0 +1,46 @@
+//go:build linux
```
rabbitstack, line 1: If the goal is to temporarily unblock the compilation, I don't have any objections to keeping the stubs that don't apply to Linux. Otherwise, we should provide only the applicable accessors.

pkg/filter/filter_windows.go
```diff
@@ -29,14 +29,14 @@ import (
 )
 
 type opts struct {
-	psnap ps.Snapshotter
+	psnap ps.Resolver
```
rabbitstack:
```suggestion
	psnap ps.Snapshotter
```

pkg/filter/filter_windows.go
```diff
@@ -29,14 +29,14 @@ import (
...
 }
 
 // Option defines the option supplied to the filter
 type Option func(o *opts)
 
-// WithPSnapshotter passes a process snapshotter reference to the filter.
-func WithPSnapshotter(psnap ps.Snapshotter) Option {
+// WithPSnapshotter passes a process resolver reference to the filter.
+func WithPSnapshotter(psnap ps.Resolver) Option {
```
rabbitstack:
```suggestion
func WithPSnapshotter(psnap ps.Snapshotter) Option {
```

pkg/pe/types_linux.go
```diff
@@ -0,0 +1,36 @@
+//go:build linux
```
rabbitstack: PE (Portable Executable) is the Windows binary format. Because of that, we shouldn't have any Linux stubs or build guardrails.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	Exe           string            `json:"exe"`
+	Cwd           string            `json:"cwd"`
+	Args          []string          `json:"args"`
+	StartTime     time.Time         `json:"started"`
+	StartBootTime uint64            `json:"start_boot_time"`
+	UID           uint32            `json:"uid"`
+	GID           uint32            `json:"gid"`
+	Username            string            `json:"username"`
+	Domain              string            `json:"domain"`
+	SID                 string            `json:"sid"`
```
rabbitstack: SID stands for Security Identifier and it it the Windows concept to represent system identities. Remove it.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	Cmdline       string            `json:"comm"`
+	Exe           string            `json:"exe"`
+	Cwd           string            `json:"cwd"`
+	Args          []string          `json:"args"`
+	StartTime     time.Time         `json:"started"`
+	StartBootTime uint64            `json:"start_boot_time"`
+	UID           uint32            `json:"uid"`
+	GID           uint32            `json:"gid"`
+	Username            string            `json:"username"`
+	Domain              string            `json:"domain"`
```
rabbitstack: I think there is no an equivalent entity to a domain in Linux.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	SID                 string            `json:"sid"`
+	SessionID           uint32            `json:"session_id"`
+	TokenIntegrityLevel string            `json:"token_integrity_level"`
```
rabbitstack: Token* and IsWow64/IsPackaged/IsProtected are Windows-only process attributes.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	TokenElevationType  string            `json:"token_elevation_type"`
+	IsTokenElevated     bool              `json:"is_token_elevated"`
+	IsWOW64             bool              `json:"is_wow_64"`
+	IsPackaged          bool              `json:"is_packaged"`
+	IsProtected         bool              `json:"is_protected"`
+	Envs                map[string]string `json:"envs"`
+	Parent              *PS               `json:"parent"`
+	Threads             map[uint32]Thread `json:"-"`
+	Mmaps               []Mmap            `json:"mmaps"`
+	Modules             []Module          `json:"modules"`
```
rabbitstack: Does not exist in Linux. Mmaps will contain the shared object mappings.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	PE                  *PE               `json:"-"`
```
rabbitstack: Does not exist in Linux. In the future, we can consider `ELF` if it proves to provide any security context.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	Tid          uint32
+	Pid          uint32
+	StartAddress va.Address
+	UstackBase   va.Address
+	UstackLimit  va.Address
+	KstackBase   va.Address
+	KstackLimit  va.Address
+}
+
+type Module struct {
```
rabbitstack: Does not apply to Linux.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	Walk(func(parent *PS) {
+		ancestors = append(ancestors, fmt.Sprintf("%s (%d)", parent.Name, parent.PID))
+	}, ps)
+	return ancestors
+}
+
+type Thread struct {
+	Tid          uint32
+	Pid          uint32
+	StartAddress va.Address
```
rabbitstack: I'm not sure if we can obtain all these attributes in Linux.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+func (ps *PS) FindMmap(addr va.Address) *Mmap {
+	for i := range ps.Mmaps {
+		if ps.Mmaps[i].BaseAddress == addr {
+			return &ps.Mmaps[i]
+		}
+	}
+	return nil
+}
+
+func (ps *PS) IsSvchost() bool {
```
rabbitstack: Windows-only. Remove.

pkg/ps/process_key.go
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ps
+
+// ProcessKey uniquely identifies a process instance across PID reuse.
+type ProcessKey struct {
```
rabbitstack: Can't see where this is being used. Anyway, I would just declare a method on the `PS` structure. In Windows, we have the `UUID` method, so it be nice if we could harmonize.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+// PE is a placeholder for PE metadata on Linux builds.
+// Linux process images are ELF; this field exists so shared
+// formatters/templates compile without Windows PE types.
+type PE struct{}
+
+func (p *PE) String() string { return "" }
+
+type PS struct {
+	sync.RWMutex
+	PID           uint32            `json:"pid"`
```
rabbitstack:
```suggestion
	PID           uint64            `json:"pid"`
```

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// PE is a placeholder for PE metadata on Linux builds.
+// Linux process images are ELF; this field exists so shared
+// formatters/templates compile without Windows PE types.
+type PE struct{}
+
+func (p *PE) String() string { return "" }
+
+type PS struct {
```
rabbitstack, line 34: Can we put doc comments on the structure + fields?

pkg/ps/types/types_linux.go
rabbitstack: This is assumed to be the main thread id.

pkg/ps/types/types_linux.go
```diff
@@ -0,0 +1,150 @@
...
+	sync.RWMutex
+	PID           uint32            `json:"pid"`
+	Ppid          uint32            `json:"ppid"`
```
rabbitstack:
```suggestion
	Ppid          uint64            `json:"ppid"`
```

pkg/ps/snapshotter_linux.go
```diff
@@ -0,0 +1,151 @@
...
+func (s *linuxSnapshotter) Put(ps *pstypes.PS) {
+	if ps == nil {
+		return
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.procs[ps.PID] = ps
+}
+
+func (s *linuxSnapshotter) UpsertSnapshot(ps *pstypes.PS) {
```
rabbitstack: What is the difference between this method and `Put`?

pkg/ps/snapshotter_linux_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"testing"
+
+	"github.com/rabbitstack/fibratus/pkg/event"
+	pstypes "github.com/rabbitstack/fibratus/pkg/ps/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLinuxSnapshotterFindPut(t *testing.T) {
```
rabbitstack:
```suggestion
func TestSnapshotterFindPut(t *testing.T) {
```

pkg/rules/compiler.go
```diff
@@ -58,12 +58,12 @@ var (
 )
 
 type compiler struct {
-	psnap     ps.Snapshotter
+	psnap     ps.Resolver
```
rabbitstack:
```suggestion
	psnap     ps.Snapshotter
```

pkg/rules/compiler.go
```diff
@@ -58,12 +58,12 @@ var (
...
 	config    *config.Config
 	approvers config.Approvers
 }
 
-func newCompiler(psnap ps.Snapshotter, cfg *config.Config) *compiler {
+func newCompiler(psnap ps.Resolver, cfg *config.Config) *compiler {
```
rabbitstack:
```suggestion
func newCompiler(psnap ps.Snapshotter, cfg *config.Config) *compiler {
```

pkg/rules/compiler_windows_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack, line 1: Can we have the tests on Linux too?

pkg/rules/engine.go
```diff
@@ -55,7 +55,7 @@ var (
 type Engine struct {
 	filters *filterset
 	config  *config.Config
-	psnap   ps.Snapshotter
+	psnap   ps.Resolver
```
rabbitstack:
```suggestion
	psnap   ps.Snapshotter
```

pkg/rules/engine.go
```diff
@@ -130,7 +130,7 @@ func (f *compiledFilter) eval(e *event.Event, valuer *filter.ValuerCache) bool {
 }
 
 // NewEngine builds a fresh rules engine instance.
-func NewEngine(psnap ps.Snapshotter, config *config.Config) *Engine {
+func NewEngine(psnap ps.Resolver, config *config.Config) *Engine {
```
rabbitstack:
```suggestion
func NewEngine(psnap ps.Snapshotter, config *config.Config) *Engine {
```

pkg/rules/engine_windows_test.go
rabbitstack, line 1: Rename to engine_windows_test.go?

pkg/rules/sequence.go
```diff
@@ -120,10 +120,10 @@ type sequenceState struct {
 	// The purpose is to enforce temporal monotonicity
 	lastMatch time.Time
 
-	psnap ps.Snapshotter
+	psnap ps.Resolver
```
rabbitstack:
```suggestion
	psnap ps.Snapshotter
```

pkg/rules/sequence.go
```diff
@@ -120,10 +120,10 @@ type sequenceState struct {
...
 }
 
-func newSequenceState(f filter.Filter, c *config.FilterConfig, psnap ps.Snapshotter) *sequenceState {
+func newSequenceState(f filter.Filter, c *config.FilterConfig, psnap ps.Resolver) *sequenceState {
```
rabbitstack:
```suggestion
func newSequenceState(f filter.Filter, c *config.FilterConfig, psnap ps.Snapshotter) *sequenceState {
```

pkg/rules/sequence_windows_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack, line 1: Use the file suffix constraint?

pkg/util/colorizer/colorizer_linux.go
rabbitstack, line 1: I'd rather rename it to colorizer_linux.go and have the noop stub.

pkg/event/bitset_linux.go
```diff
@@ -0,0 +1,23 @@
+//go:build linux
```
rabbitstack, line 1: This is never going to be a thing in Linux. Remove it ;)

pkg/util/log/logger.go
```diff
@@ -1,3 +1,5 @@
+//go:build windows
```
rabbitstack: Rename the file to `logger_windows.go` and remove the build tag?

pkg/util/ntstatus/status_linux.go
```diff
@@ -0,0 +1,35 @@
+//go:build linux
```
rabbitstack: This is never going to be a thing in Linux. Remove it ;)

internal/bootstrap/bootstrap_linux.go
mostafa: Yes. The Linux App stubs are intentional for this PR; capture wiring and fuller bootstrap behavior will land in follow-up PRs. run_linux.go already consumes bootstrap.NewApp the same way as Windows (minus Windows-only options like debug privilege).

pkg/alertsender/alert_linux.go
rabbitstack, line 1: I'd refactor to `alert_linux.go`. So in the future it can accommodate more stuff than just process types.

pkg/alertsender/alert_windows.go
rabbitstack, line 1: I'd refactor to `alert_windows.go`. So in the future, it can accommodate more stuff than just process types.

pkg/api/listener_linux.go
```diff
@@ -0,0 +1,34 @@
...
+package api
+
+import (
+	"context"
+	"fmt"
+	"net"
+)
+
+// DialPipe is unavailable on Linux. The HTTP API uses TCP transports instead.
+func DialPipe(pipePath string) func(context.Context, string, string) (net.Conn, error) {
```
rabbitstack: I'd factor this out. Something like:

```go
func DialLocalTransport(path string) func(context.Context, string, string) (net.Conn, error)
```

On Windows, it would remain as is (building the named pipe connection), while on Linux it would produce the UNIX domain socket transport (`net.Dial("unix", path)`)

pkg/config/alertsender_linux.go
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package config
+
+import "github.com/rabbitstack/fibratus/pkg/alertsender"
+
+func (c *Config) loadPlatformAlertSender(typ string, config interface{}, configs *[]alertsender.Config) error {
```
rabbitstack:
```suggestion
func (c *Config) loadPlatformAlertSender(_ string, _ interface{}, _ *[]alertsender.Config) error {
```

pkg/config/alertsender_linux.go
```diff
@@ -0,0 +1,31 @@
...
+	// Platform-specific Linux alert senders can be loaded here.
+	_ = typ
```
rabbitstack:
```suggestion
```

pkg/config/alertsender_linux.go
```diff
@@ -0,0 +1,31 @@
...
+	_ = config
```
rabbitstack:
```suggestion
```

pkg/config/alertsender_linux.go
```diff
@@ -0,0 +1,31 @@
...
+	_ = configs
```
rabbitstack:
```suggestion
```

pkg/config/config.go
```diff
@@ -0,0 +1,258 @@
...
+	capFile     = "cap.file"
+	configFile  = "config-file"
+	forwardMode = "forward"
+)
+
+// Config stores configuration options for fine-tuning the behaviour of Fibratus.
+type Config struct {
+	platformConfig
+
+	EventSource  EventSourceConfig `json:"eventsource" yaml:"eventsource"`
```
rabbitstack: Can we keep the doc comments from the previous structure definition?

pkg/config/config.go
```diff
@@ -0,0 +1,258 @@
...
+	run      bool
+	list     bool
+	stats    bool
+	validate bool
+}
+
+// Option configures the command-specific configuration surface.
+type Option func(*Options)
+
+func WithCapture() Option  { return func(o *Options) { o.capture = true } }
```
rabbitstack: Can we keep the doc comments?

pkg/config/config.schema.json
```diff
@@ -335,6 +335,12 @@
     "eventsource": {
       "type": "object",
       "properties": {
+        "enable-process": {
```
rabbitstack: Controlling the emission of process events was deliberately omitted as they build the core of the system and wouldn't make much sense inhibiting them.

pkg/config/config.schema.json
```diff
@@ -335,6 +335,12 @@
...
+          "type": "boolean"
+        },
+        "enable-file": {
```
rabbitstack: Is this the same as `enable-fileio`?

pkg/config/config.schema.json
rabbitstack, line 1: I have a gut feeling we'll need per-platform schema separation. I'll take care of that ;)

pkg/event/params/names_windows.go
```diff
@@ -0,0 +1,91 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package params
+
+const (
+	NTStatus                   = "status"
```
rabbitstack: Can we retain the doc comments?

pkg/event/params/param_types.go
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+package params
+
+// Type defines an event parameter type.
+type Type uint16
+
+const (
+	Null    Type = 0
```
rabbitstack: Can we retain the doc comments?

pkg/filter/fields/defs_linux.go
```diff
@@ -0,0 +1,36 @@
...
+
+import "github.com/rabbitstack/fibratus/pkg/event/params"
+
+func platformString() params.Type     { return params.String }
+func platformAnsiString() params.Type { return params.String }
+
+var fields = platformFields(map[Field]FieldInfo{
+	EvtPID:      {EvtPID, "process identifier generating the event", params.Uint64, []string{"evt.pid = 6"}, nil, nil},
+	EvtTID:      {EvtTID, "thread identifier generating the event", params.Uint64, []string{"evt.tid = 1024"}, nil, nil},
+	KevtPID:     {KevtPID, "process identifier generating the event", params.Uint64, []string{"kevt.pid = 6"}, &Deprecation{Since: "3.0.0", Fields: []Field{EvtPID}}, nil},
```
rabbitstack: `kevt.*` fields are deprecated. We can remove them.

pkg/ps/snapshotter.go
rabbitstack, line 1: Rename to `snapshotter_windows.go`

pkg/ps/snapshotter_mock_windows.go
rabbitstack, line 1: Rename to `snapshotter_mock_windows.go`

pkg/util/ntstatus/status.go
rabbitstack, line 1: Rename back to `status.go` because the constraint is dictated by the build tag.

rabbitstack (changes requested):

rabbitstack (approved):
LGTM

PR #729, opened 2026-09-14
feat(ebpf): Add Linux process event source

mostafa:

Adds the live Linux eBPF process source so `fibratus run` can capture `execve`, `exit`, and `clone` instead of stopping at an unwired capture stub. The source loads committed CO-RE objects, shares maps across separately generated collections, snapshots thread-group leaders with `iter/task`, and reconciles hot events against that baseline using PID plus start boot time.

Successful clones take child identity from `sched_process_fork`, and process exits are emitted from `sched_process_exit` because `exit`/`exit_group` never return (their syscall exit tracepoints never fire) and signal-terminated processes never invoke them. Filename bytes for `execve`/`execveat` are copied into a scratch map on enter so a successful exec cannot reread a replaced address space. Executable path and cmdline are best-effort `/proc/<pid>/{exe,cmdline}` reads only. Missing kernel 5.9, usable `/sys/kernel/btf/vmlinux`, or required program/map helpers fail the open path with a concrete error.

---

/kind feature

---

/area instrumentation

/area telemetry

/area tests

/area ci

/area build

---

- x86_64 only. The loader rejects other GOARCH values.
- Generated `bpf2go` bindings and `.o` objects are committed. Ordinary Linux builds do not need clang. `internal/ebpf/generate.sh` regenerates them.
- Shared maps (`events`, `drop_count`, `scratch`) are owned by the execve collection and injected into exit/clone/iterator collections with `MapReplacements`.
- Startup sequence: load maps, start the ring buffer reader, attach the hot path, queue inbound events, run `iter/task`, upsert baseline process state, replay pending events in ring-buffer order, then switch to live dispatch.
- Credential parameters carry effective uid/gid.
- Process events are always captured. File/net/memory families are not in this change.
- Linux bootstrap now constructs the process snapshotter, optional rule engine, aggregator, local API (TCP or `unix://`), and signal handling. Filaments remain unsupported.
- Capabilities expected at runtime: `CAP_BPF`, `CAP_PERFMON`, and `CAP_SYS_PTRACE` for incidental `/proc` enrichment.
- Verified:
  - Docker Linux: `go test ./internal/ebpf ./internal/bootstrap ./pkg/event ./pkg/ps ./pkg/api ./pkg/util/signals`
  - Docker Linux: `go build -o /tmp/fibratus ./cmd/fibratus`
  - `GOOS=windows GOARCH=amd64 go test -exec=true ./pkg/event ./pkg/util/signals ./pkg/api ./internal/bootstrap`

---

Yes. On a 5.9+ kernel with runtime BTF, `fibratus run` opens the eBPF process source and emits enriched `execve`, `exit`, and `clone` events into the existing aggregator and outputs path.


internal/ebpf/c/common/events.h
```diff
@@ -0,0 +1,125 @@
...
+struct trace_event_raw_sys_exit {
+	unsigned short common_type;
+	unsigned char common_flags;
+	unsigned char common_preempt_count;
+	int common_pid;
+	long id;
+	long ret;
+};
+
+enum fibratus_event_kind {
```
rabbitstack:
```suggestion
enum syscall_event_kind {
```

internal/ebpf/c/common/events.h
```diff
@@ -0,0 +1,125 @@
...
+	u32 pid;
+	u32 tid;
+	u32 tgid;
+	u32 ppid;
+	u32 uid;
+	u32 gid;
+	u32 syscall_id;
+	u32 pad;
+	s64 retval;
+	u64 clone_flags;
```
rabbitstack: I assume once you extend the telemetry spectrum to support other than process events, `clone_flags` would be a bit misleading, right?

internal/ebpf/c/common/events.h
```diff
@@ -0,0 +1,125 @@
...
+	EVT_KIND_CLONE = 3,
+};
+
+/* Stable semantic type IDs. Keep in sync with pkg/event.Type on Linux. */
+#define EVT_TYPE_UNKNOWN 0
+#define EVT_TYPE_EXECVE 1
+#define EVT_TYPE_EXIT 2
+#define EVT_TYPE_CLONE 3
+
+struct fibratus_event {
```
rabbitstack:
```suggestion
struct syscall_event {
```

internal/ebpf/c/clone.bpf.c
```diff
@@ -0,0 +1,166 @@
...
+	u64 key;
+
+	if (!child || !parent)
+		return 0;
+
+	e = reserve_event();
+	if (!e)
+		return 0;
+
+	e->kind = EVT_KIND_CLONE;
```
rabbitstack: Do kind and type designate the same event topology?

internal/ebpf/loader.go
```diff
@@ -0,0 +1,254 @@
...
+		return 0
+	}
+	_ = l.exec.DropCount.Lookup(&key, &drop)
+	return drop
+}
+
+// syscallsGroup is the tracefs group hosting raw syscall tracepoints.
+const syscallsGroup = "syscalls"
+
+func (l *loader) attachHotPath() error {
```
rabbitstack: Out of curiosity, what are "hot path" tps?

rabbitstack (approved):
Overall, looks pretty solid. Just a few naming nits and some questions. Approving...

internal/ebpf/c/common/events.h
mostafa: Good point, and it went one step further in 4c2ae54: the enum is gone entirely. The kind always mirrored the type, so the record now carries only the stable type ID.

internal/ebpf/c/common/events.h
mostafa: Applied in 4c2ae54.

internal/ebpf/c/common/events.h
mostafa: Right. Renamed it to a per-type `flags` field in 4c2ae54, with clone storing the raw clone flags. Future event families can reuse it for their own flag words without another layout change.

internal/ebpf/c/clone.bpf.c
mostafa: They effectively did, which is why kind is now removed in 4c2ae54. The only extra information it carried was distinguishing iter/task snapshot records, and those are now type 0: that value maps to the unknown type in Go, so a snapshot record can never be dispatched as a live event.

internal/ebpf/loader.go
mostafa: They are the continuously firing live-capture programs (raw syscall tracepoints plus the sched_process_fork/exit tp_btf hooks) as opposed to the one-shot iter/task snapshot program. Since the name was not self-explanatory, 4c2ae54 renames it to attachPrograms and adds a comment, including the ordering contract that the ring buffer reader starts before attachment so no events are lost.

PR #730, opened 2026-09-15
feat(ebpf): Add file, network, memory, and process-control syscall telemetry

mostafa:

Expands the Linux eBPF source beyond the process vertical slice to the MVP syscall matrix: `openat` (with `open`/`openat2`), `unlink`/`unlinkat`, `rename`/`renameat`/`renameat2`, `connect`, `accept`/`accept4`, `mmap`, `process_vm_readv`/`process_vm_writev`, `kill`/`tkill`/`tgkill`, `ptrace`, and `prctl`. Each family lives in its own BPF program object sharing the ring buffer, scratch, and enabled-type maps through `MapReplacements`.

The shared `syscall_event` record grows generic argument slots, a second path buffer, and truncation bits. Enter-state is staged in a per-CPU heap array (two 256-byte paths exceed the BPF stack) and correlated to the exit through a global LRU hash keyed by pid_tgid: the task can migrate CPUs mid-syscall, so the correlation map cannot be per-CPU, and LRU eviction reclaims entries whose exit never fires. Sockaddr copies clamp to the caller's addrlen because a fixed-size read past the buffer can cross into an unmapped page and fail wholesale; accept reads the value-result addrlen and peer address at exit. Emission is gated in the kernel by an enabled-type array populated from the event source config, so disabled families cost neither ring buffer traffic nor userspace decode.

Successful `mmap` calls update the process state with the new mapping (base, size, protection, anonymous or file-backed kind).

---

/kind feature

---

/area instrumentation

/area telemetry

/area tests

---

- Stable semantic type IDs 4-14 extend the Linux catalog; the raw syscall number stays an event parameter, so `openat`/`openat2`/`open` all map to one `openat` event.
- Process-control events (`kill`, `ptrace`, `prctl`) are always captured with the process family. File, network, and memory families follow `enable-fileio`, `enable-net`, and `enable-mem`, enforced twice: tracepoints are not attached, and the kernel-side enabled map stays zero.
- Only legacy syscall tracepoints (`open`, `unlink`, `rename`, `tkill`, and the existing `fork`/`vfork`) are optional attachments, following the earlier finding that some kernels refuse perf links on them. Modern variants (`openat2`, `renameat2`, `accept4`, `tgkill`) must attach.
- Scratch entries are deleted on every exit path, including ring-buffer-full reserve failures.
- Objects regenerated with Ubuntu 24.04 clang 18 to match the drift check.
- Verified:
  - Docker Linux: `./internal/ebpf/generate.sh` followed by a clean `git diff -- internal/ebpf`
  - Docker Linux: `go test ./internal/ebpf/... ./pkg/event/...`
  - Matrix test asserts every catalog type decodes its documented parameters; live integration test covers `openat`, `rename`, `unlink`, and `kill` end to end.

---

Yes. On a 5.9+ kernel with runtime BTF, `fibratus run` now emits file, network, memory, and process-control syscall events alongside process telemetry, honoring the `enable-fileio`, `enable-net`, and `enable-mem` event source settings.


internal/ebpf/c/common/events.h
```diff
@@ -4,12 +4,24 @@
...
+#include "bpf_core_read.h"
 
 #ifndef BPF_ANY
 #define BPF_ANY 0
 #endif
 
 #define EVT_FILENAME_LEN 256
+#define EVT_TYPE_MAX 32
+
+#define TRUNC_FILENAME  (1u << 0)
```
rabbitstack: Could we document these two and perhaps find a better name for `TRUNC_FILENAME2`?

internal/ebpf/c/common/events.h
```diff
@@ -48,18 +71,31 @@ struct syscall_event {
...
-	/* Interpreted per event type; clone stores the raw clone flags. */
+	/* Interpreted per event type: clone flags, open flags, mmap flags, accept flags. */
 	u64 flags;
+	u64 arg0;
+	u64 arg1;
+	u64 arg2;
+	u64 arg3;
 	u64 start_boottime;
 	u64 timestamp_ns;
+	u32 truncated;
```
rabbitstack, line 84: What is the purpose of this field?

internal/ebpf/c/common/events.h
```diff
@@ -48,18 +71,31 @@ struct syscall_event {
...
+	u32 pad;
 	u8 comm[TASK_COMM_LEN];
 	u8 filename[EVT_FILENAME_LEN];
+	u8 filename2[EVT_FILENAME_LEN];
```
rabbitstack: If possible, rename here and in `scratch_value` so it better conveys the purpose.

internal/ebpf/c/ctl.bpf.c
```diff
@@ -0,0 +1,106 @@
+//go:build ignore
+
+#include "common/events.h"
+#include "bpf_tracing.h"
+
+char LICENSE[] SEC("license") = "Dual MIT/GPL";
+
+static __always_inline int handle_kill_enter(u64 pid, u64 sig)
```
rabbitstack: General question. Do we need to handle the syscall enter hook at all? Check [this](https://github.com/falcosecurity/libs/blob/master/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/kill.bpf.c). The pattern is equally present across all other programs, because `exit` has access to all syscall args + return value.
That would simplify and speed up the probe code.

internal/ebpf/c/common/events.h
```diff
@@ -48,18 +71,31 @@ struct syscall_event {
 	u32 gid;
 	u32 syscall_id;
 	s64 retval;
-	/* Interpreted per event type; clone stores the raw clone flags. */
+	/* Interpreted per event type: clone flags, open flags, mmap flags, accept flags. */
 	u64 flags;
+	u64 arg0;
```
rabbitstack, line 78: Do you think it is worth separating the event payload per type, e.g. `clone_event`, `open_event`, etc? We don't have to do it now, but I think it would improve the readability and future maintainability (every event representation contains typed fields instead of opaque args).

internal/ebpf/consumer.go
```diff
@@ -62,6 +63,17 @@ func applyProcessState(psnap ps.Snapshotter, evt *event.Event) {
 			evt.PS = existing
 			_ = psnap.AddThread(evt)
 		}
+	case evt.Type == event.Mmap && succeeded(evt):
+		ok, existing := psnap.Find(evt.PID)
+		if ok && existing != nil {
+			evt.PS = existing
+			existing.AddMmap(pstypes.Mmap{
```
rabbitstack: We're adding but never purging the mmaped region? Do you think it would be better to just retain file-backed mmaps or everything?

internal/ebpf/loader.go
```diff
@@ -120,6 +190,45 @@ func (l *loader) dropCount() uint64 {
...
+			val = 1
+		}
+		if err := l.exec.Enabled.Put(&key, &val); err != nil {
+			return fmt.Errorf("enabling %s: %w", typ, err)
+		}
+	}
+	return nil
+}
+
+func enabledTypes(cfg *config.EventSourceConfig) map[event.Type]bool {
```
rabbitstack, line 210: Note for the future. A bitmask is generally more performant than a map lookup.

rabbitstack (changes requested):

internal/ebpf/c/common/events.h
mostafa: Documented both bits and renamed `TRUNC_FILENAME2` to `TRUNC_AUX` to match the buffer it refers to. `TRUNC_FILENAME` is `filename[]` (exec path, open/unlink/rename source); `TRUNC_AUX` is `aux[]` (rename destination or sockaddr bytes).

internal/ebpf/c/common/events.h
mostafa: Explicit padding so `comm[]` starts on an 8-byte boundary. It has no semantic value. Commented in place.

internal/ebpf/c/common/events.h
mostafa: Renamed to `aux` in `syscall_event` and `scratch_value`, with comments on both buffers. `filename[]` is the primary path; `aux[]` is the rename destination or the sockaddr blob. Typed per-event payloads would make this go away, as noted below.

internal/ebpf/c/ctl.bpf.c
mostafa: Yes for register-only syscalls. The per-syscall `sys_exit_*` tracepoint only has `id` and `ret`, which is why enter+scratch existed. Falco's pattern works because `tp_btf/sys_exit` supplies `pt_regs`. kill/ptrace/prctl now attach there, read args from the registers (arg3 is r10, not rcx), and drop the enter probes.

Enter stays for execve (successful exec replaces the address space) and clone (flags have to survive until `sched_process_fork`). File/net still copy userspace strings through scratch on the per-syscall tracepoints; folding those onto this `sys_exit` dispatcher is the next step so we do not attach a second global exit program for mmap.

internal/ebpf/c/common/events.h
mostafa: Agreed, and it would make `arg0`/`aux` unnecessary. Kept one shared ringbuf layout for this slice so every family still goes through the same decode path. Happy to split `clone_event`/`open_event`/etc. once the catalog is denser.

internal/ebpf/consumer.go
mostafa: File-backed only. Anonymous mappings are frequent (heap, stacks, JIT) and we do not hook munmap yet, so keeping them would grow without bound. Process state now stores mappings with `MAP_ANONYMOUS` clear and `fd >= 0`. The mmap event itself is still emitted either way. munmap can call the existing `RemoveMmap` once that syscall is in the catalog.

internal/ebpf/loader.go
mostafa: Agreed. The array lookup is one helper call per event and is fine at this catalog size. Will switch to a bitmask when we next touch the enabled path, likely with the shared `sys_exit` dispatcher.

rabbitstack (approved):

PR #731, opened 2026-09-16
feat(filter): Add Linux filter fields and accessors

mostafa:

Makes the syscall telemetry introduced on `linux-port` filterable. The Linux field catalog grows from five identity fields to the full event matrix, and real accessors replace the noop stubs for the file, network, memory, and thread families.

Shared field names are reused only where the semantics match the Windows counterpart: `file.path`, `file.name`, `file.extension`, `file.path.stem`, `net.dip`/`net.sip`/`net.dport`/`net.sport`, `mem.address`/`mem.size`/`mem.protection`, `ps.uuid`, and the parent process fields. Windows-specific fields (`mem.type`, `mem.alloc`, `registry.*`, `pe.*`, `handle.*`, `dns.*`, and the deprecated `kevt.*` aliases) do not enter the Linux catalog and fail to compile in a filter expression. Linux-native concepts get their own names: `ps.uid`/`ps.gid`, `ps.signal`, `ps.target.pid`, `ps.ptrace.request`, `ps.prctl.option`, `ps.clone.flags`, `file.new_path`/`file.dirfd`/`file.fd`/`file.flags`/`file.mode`/`file.truncated`, `net.family`/`net.path`/`net.fd`, `mem.flags`/`mem.fd`/`mem.offset`/`mem.target.pid`, `thread.tid`/`thread.pid`, and `evt.retval`/`evt.syscall`/`evt.truncated`.

The deprecated `kevt.*` aliases move from the common catalog into the Windows platform catalog, so Windows behavior is unchanged while new Linux filters use the supported `evt.*` names from the start.

---

/kind feature

---

/area filters

/area tests

---

- File, network, and memory accessors follow `enable-fileio`, `enable-net`, and `enable-mem`, mirroring how the Windows constructor gates its accessors. The thread accessor is always on and only answers for thread-creating clones.
- The evt accessor resolves `evt.retval`, `evt.syscall`, and `evt.truncated` through a platform hook because the backing parameter names exist only on Linux; the Windows hook is a noop, so the shared accessor compiles on both platforms.
- `file.truncated` reports either truncation bit for file events, since the aux buffer holds the rename destination path there.
- `mem.protection`, `mem.flags`, and `file.flags`/`file.mode` expose the raw bitmask/mode values captured by the eBPF programs; symbolic rendering can layer on top later without changing field identity.
- Verified:
  - Docker Linux: `go test ./pkg/filter/... ./pkg/rules/`
  - Windows cross-compilation: `GOOS=windows go build ./pkg/...` and `go vet ./pkg/filter/...`
  - Tests assert compile and evaluation for every Linux event type, shared-name semantics, missing-value defaults, truncation bits, and that Windows-only and `kevt.*` fields are rejected at compile time.

---

Yes. On Linux, filter expressions and rules can now reference process, file, network, memory, and thread fields for all captured syscall events, for example `fibratus run evt.name = 'connect' and net.dport = 443`.


pkg/filter/fields/defs.go
```diff
@@ -117,6 +117,20 @@ const (
...
+	// PsSignal represents the POSIX signal number
+	PsSignal Field = "ps.signal"
+	// PsTargetPID represents the target process identifier of kill, ptrace, or process_vm operations
+	PsTargetPID Field = "ps.target.pid"
+	// PsPtraceRequest represents the ptrace request code
+	PsPtraceRequest Field = "ps.ptrace.request"
+	// PsPrctlOption represents the prctl option code
+	PsPrctlOption Field = "ps.prctl.option"
+	// PsCloneFlags represents the clone flags bitmask
+	PsCloneFlags Field = "ps.clone.flags"
```
rabbitstack:
```suggestion
	PsCloneFlags Field = "ps.clone_flags"
```

pkg/filter/fields/defs.go
```diff
@@ -361,6 +377,12 @@ const (
 	EvtNparams Field = "evt.nparams"
 	// EvtArg represents the field sequence for generic argument access
 	EvtArg Field = "evt.arg"
+	// EvtRetval represents the syscall return value
+	EvtRetval Field = "evt.retval"
+	// EvtSyscall represents the raw architecture-specific syscall number
+	EvtSyscall Field = "evt.syscall"
```
rabbitstack: What do you think about `evt.syscall_id`?

pkg/filter/fields/defs.go
```diff
@@ -361,6 +377,12 @@ const (
...
+	// EvtTruncated indicates whether variable-length event fields were truncated
+	EvtTruncated Field = "evt.truncated"
```
rabbitstack: Any foreseeable use case for this field in the context of detection rules? I'd keep this internal/hidden. Even so, if someone needs it, they could use a generic event accessor (`evt.arg[]`).

pkg/filter/fields/defs.go
```diff
@@ -442,6 +464,12 @@ const (
 	NetSIPNames Field = "net.sip.names"
 	// NetDIPNames represents the destination IP names
 	NetDIPNames Field = "net.dip.names"
+	// NetFamily represents the socket address family
+	NetFamily Field = "net.family"
+	// NetPath represents the filesystem path of an AF_UNIX socket
+	NetPath Field = "net.path"
```
rabbitstack:
```suggestion
	NetUnixPath Field = "net.unix_path"
```

pkg/filter/fields/defs.go
```diff
@@ -453,6 +481,18 @@ const (
 	FilePathStem Field = "file.path.stem"
 	// FileExtension represents the file extension (e.g. .exe or .dll)
 	FileExtension Field = "file.extension"
+	// FileNewPath represents the destination path of a rename
+	FileNewPath Field = "file.new_path"
```
rabbitstack: To repurpose for future event types that deal with source/target file paths (mv, cp, etc.):
```suggestion
	FilePathTarget Field = "file.path.target"
```

pkg/filter/fields/defs.go
```diff
@@ -453,6 +481,18 @@ const (
...
+	// FileDirFD represents a directory file descriptor, including AT_FDCWD
+	FileDirFD Field = "file.dirfd"
+	// FileFD represents a file descriptor
+	FileFD Field = "file.fd"
+	// FileFlags represents the raw openat, unlinkat, or renameat flags bitmask
+	FileFlags Field = "file.flags"
+	// FileMode represents the file creation mode supplied to openat
+	FileMode Field = "file.mode"
+	// FileTruncated indicates whether the file path was truncated
+	FileTruncated Field = "file.truncated"
```
rabbitstack: Unlikely to be used in detection rules. If ever needed, rule authors can access it via the generic event param accessor.

pkg/filter/fields/defs.go
```diff
@@ -639,6 +679,14 @@ const (
 	MemProtection Field = "mem.protection"
 	// MemProtectionMask identifies the field that represents the memory protection in mask notation
 	MemProtectionMask Field = "mem.protection.mask"
+	// MemFlags represents the mmap flags bitmask
```
rabbitstack: Do you think it is worth prepending the `mmap` prefix to describe precisely the value this field targets and at the same time avoid ambiguity? 

```
MemMmapFlags = "mem.mmap.flags"
MemMmapFD = "mem.mmap.fd"
MemMmapOffset = "mem.mmap.offset"
```

rabbitstack (approved):
LGTM, after addressing the changes!

pkg/filter/fields/defs.go
mostafa: Renamed to `ps.clone_flags`.

pkg/filter/fields/defs.go
mostafa: Renamed to `evt.syscall_id` (`EvtSyscallID`).

pkg/filter/fields/defs.go
mostafa: Dropped `evt.truncated` from the catalog. Truncation stays on the `truncated` event parameter and is reachable with `evt.arg[truncated]`.

pkg/filter/fields/defs.go
mostafa: Renamed to `net.unix_path` (`NetUnixPath`).

pkg/filter/fields/defs.go
mostafa: Renamed to `file.path.target` (`FilePathTarget`).

pkg/filter/fields/defs.go
mostafa: Dropped `file.truncated` the same way as `evt.truncated`. Use `evt.arg[truncated]` for the bitmask.

pkg/filter/fields/defs.go
mostafa: Prefixed the mmap fields: `mem.mmap.flags`, `mem.mmap.fd`, and `mem.mmap.offset`.

PR #732, opened 2026-09-16
feat(build,linux): Add Makefile script for build targets

rabbitstack:

The Makefile script automates the process of compiling the eBPF programs, Fibratus binary, running tests, formatting, etc. The `bpf2go` now drops all generated Go files + ebpf bytecode inside a separate bpf package, making it easier to navigate and distinguish the generated code.

---

> Uncomment one or more `/kind <>` lines:

> /kind feature (non-breaking change which adds functionality)

> /kind bug-fix (non-breaking change which fixes an issue)

> /kind refactor (non-breaking change that restructures the code, while not changing the original functionality)

> /kind breaking (fix or feature that would cause existing functionality to not work as expected

> /kind cleanup

/kind improvement

> /kind design

> /kind documentation

> /kind other (change that doesn't pertain to any of the above categories)

---

> Uncomment one or more `/area <>` lines:

> /area instrumentation

> /area telemetry

> /area rule-engine

> /area filters

> /area yara

> /area event

> /area captures

> /area alertsenders

> /area outputs

> /area rules

> /area filaments

> /area config

> /area cli

> /area tests

> /area ci

/area build

> /area docs

> /area deps

> /area evasion

> /area other

---

---


Makefile
```diff
@@ -0,0 +1,58 @@
...
+	-O2 \
+	-g \
+	-D__TARGET_ARCH_x86
+
+# Find all eBPF sources.
+BPF_SOURCES := $(wildcard $(BPF_DIR)/c/*.bpf.c)
+BPF_NAMES := $(patsubst $(BPF_DIR)/c/%.bpf.c,%,$(BPF_SOURCES))
+BPF_TARGETS := $(foreach name,$(BPF_NAMES), \
+	$(BPF_DIR)/$(name)_bpfel.go \
+	$(BPF_DIR)/$(name)_bpfeb.go)
```
mostafa: These targets are `internal/ebpf/$(name)_bpfel.go`, but bpf2go writes to `internal/ebpf/bpf/` via `-output-dir`. Make never sees the generated files, so `make ebpf` rebuilds on every invocation. Point `BPF_TARGETS` and the pattern rule at `$(BPF_DIR)/bpf/`.

Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+BPF_FLAGS := \
+	-go-package bpf \
+	-output-dir $(BPF_DIR)/bpf \
+	-cc clang \
+	-target bpfel,bpfeb \
+	-tags linux
+
+BPF_CFLAGS := \
+	-I./$(BPF_DIR)/spike/c \
```
mostafa: This puts `spike/c` ahead of the production headers. The previous generator searched `c`, then `c/common`, then `spike/c` so spike `vmlinux.h` is only a fallback. Reverse the include order.

Makefile
```diff
@@ -0,0 +1,58 @@
...
+# Specifies a list of build flags
+TAGS ?= ""
+
+BPF_DIR := internal/ebpf
+
+BPF_FLAGS := \
+	-go-package bpf \
+	-output-dir $(BPF_DIR)/bpf \
+	-cc clang \
+	-target bpfel,bpfeb \
```
mostafa, line 23: `CLANG` is defined later but unused; `-cc clang` is hardcoded. Use `-cc $(CLANG)` and define `CLANG` before `BPF_FLAGS` because this assignment is immediate (`:=`).

Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+CLANG ?= clang
+GO ?= go
+GOFMT ?= gofmt
+BPF2GO ?= go run github.com/cilium/ebpf/cmd/bpf2go@$(BPF2GO_VERSION)
+
+.PHONY: ebpf
+ebpf: $(BPF_TARGETS)
+$(BPF_DIR)/%_bpfel.go $(BPF_DIR)/%_bpfeb.go &: $(BPF_DIR)/c/%.bpf.c
+	$(BPF2GO) $(BPF_FLAGS) -output-stem $* $(shell echo $* | sed 's/^./\U&/') $(BPF_DIR)/c/$*.bpf.c -- $(BPF_CFLAGS)
```
mostafa: `sed 's/^./\U&/'` is GNU sed. `awk '{print toupper(substr($0,1,1)) substr($0,2)}'` produces the exported bpf2go identifier without that dependency.

Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+.PHONY: build
+build:
+	$(GO) build -tags $(TAGS) -o ./cmd/fibratus/fibratus ./cmd/fibratus/
+
+.PHONY: fmt
+fmt:
+	$(GOFMT) -e -s -l -w pkg cmd
```
mostafa: `fmt` skips `internal/`, which is where the Linux source lives.

Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+.PHONY: test
+test:
+	$(GO) test ./...
```
mostafa: `go test ./...` fails on Linux (Windows-only cmdline/utf16/yara tests). The linux-port job already lists the packages that pass; `make test` should use that set.

.github/workflows/linux-port.yml
```diff
@@ -35,9 +35,8 @@ jobs:
           clang --version
       - name: Check generation drift
         run: |
-          chmod +x internal/ebpf/generate.sh
-          ./internal/ebpf/generate.sh
-          git diff --exit-code -- internal/ebpf
+          make ebpf
```
mostafa: After the output-dir mismatch is fixed, `make ebpf` becomes incremental and a fresh checkout can skip regeneration when timestamps are equal. Use `make -B ebpf` so the drift check always rebuilds.

mostafa (changes requested):
The bpf package split is the right direction. Makefile target paths do not match `-output-dir`, so `make ebpf` always rebuilds, and the committed objects fail the Ubuntu clang 18 drift check.

Proposed fix with regenerated objects: https://github.com/rabbitstack/fibratus/pull/734

